PR #646, opened 2021-10-06
Enable cargo resolver v2

CAD97:
Primarily to see if this catches #645 in CI.


CAD97:
@googlebot I signed it!

PR #825, opened 2022-02-22
ignore `test_abstract_private` on beta channel

philipcraig:
Fixes https://github.com/google/autocxx/pull/823

Relates to #818


philipcraig:
not needed as beta channel is fixed

PR #871, opened 2022-03-03
Move away from deprecated C++ APIs in the demo

bsilver8192:
A bit less cruft printed during tests which use it.


bsilver8192:
CI failure looks like flaky external dependencies, I don't see how to rerun it.

PR #1147, opened 2022-08-14
Support multiple AUTOCXX_RS_JSON_ARCHIVE entries

bsilver8192:
This is helpful when using relative paths with doctests, because
doctests are compiled from a different working directory than the crate
itself. Bazel requires relative paths in environment variables so that
caching works across machines, for example.


bsilver8192:
This is a draft due to depending on #1148, and GitHub's lack of good support for dependent PRs. Feel free to mark as in progress (I think anybody can do that?) and merge once that's merged.

PR #1260, opened 2023-03-13
Fix soundness issue due to moveit DerefMove

silvanshade:
This is a companion PR to https://github.com/google/moveit/pull/37.

Some background on this:

I recently started working on some rather complicated bindings to parts of LLVM and was originally using plain `cxx` but decided to try `autocxx` to hopefully save some time and generate more complete bindings.

The new machinery in `autocxx` relating to constructors seems very interesting and powerful, but I noticed (when I was trying to understand originally how to use it), that I was able to do some suspicious things, like create a new instance of some unpin type `T: !Unpin` as a `Pin<MoveRef<T>>`, but then actually get a `T` back out with a combination of `&move *val` followed by `MakeRef::into_inner`.

This didn't seem like it should be possible without `unsafe` and after a bit of searching I saw https://github.com/google/moveit/issues/34.

The associated PR in the `moveit` repo fixes the issue but slightly changes the interface to `mov`, hence this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -961,8 +961,7 @@ dependencies = [
 [[package]]
 name = "moveit"
 version = "0.5.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7d756ffe4e38013507d35bf726a93fcdae2cae043ab5ce477f13857a335030d"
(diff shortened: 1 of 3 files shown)


silvanshade:
Note that the associated PR for this is now https://github.com/google/moveit/pull/38. I will submit an updated PR whenever that issue is resolved.

PR #1426, opened 2025-01-20
[DRAFT] Feature/throws exception

### LINKED ISSUE #1412: Feature Request: Exception Support
First off, thank you for the excellent tooling you provide! I really appreciate `autocxx`, and it made it much easier to release [torrancew/xapian-rs](https://github.com/torrancew/xapian-rs).

In the process of writing the bindings, I ran into a few pain points, but the most difficult one was the lack of support for exceptions in `autocxx`. Upstream Xapian code `throw`s quite a bit, and it would be great to express most of those in the Rust bindings without resorting to writing almost everything by hand.

Would you be open to receiving patches for this? If so, could you provide some high level guidance on how best to start? I'm fairly comfortable in Rust, but my C++ knowledge is... dated. I know enough to know that I don't know much, but I can usually get something that compiles thrown together -- but I'm willing to iterate if you're open to the patches and some potentially naive lines of questioning.

Thanks again!

torrancew:
Fixes #1412 

_This is an early draft to solicit feedback. It is functional, but there are a couple of implementation details that felt rather clumsy:_
- _Mapping a user specified qualified name (eg `Xapian::Database::add_document`) to the mangled representations available during function analysis_
- _Representing the transformations themselves within the existing machinery (the throwable bit is more or less just bolted on at this phase)_

_And of course some obvious missing bits like documentation, additional test cases, etc. Let me know what you think about the current approach, and anything you'd like to see improved -- I'm happy to iterate!_

This patch introduces a `throws!` macro, so that individual functions and methods can be marked as throwing an exception. Example usage can be found in the provided unit tests, but generally resembles:

```rust
import_cpp! {
    #include "foo.h"
    generate!("do_something")
    throws!("do_something")
}
```

- [x] CLA signed
- [x] Tests pass

### DIFF
--- a/engine/src/conversion/analysis/fun/mod.rs
+++ b/engine/src/conversion/analysis/fun/mod.rs
@@ -167,6 +167,8 @@ pub(crate) struct FnAnalysis {
     /// subclass entries for them. But we do not want to have them
     /// be externally callable.
     pub(crate) ignore_reason: Result<(), ConvertErrorWithContext>,
+    /// Whether this method can throw an exception
+    pub(crate) throwable: bool,
     /// Whether this can be called by external code. Not so for
     /// protected methods.
     pub(crate) externally_callable: bool,
@@ -825,7 +827,7 @@ impl<'a> FnAnalyzer<'a> {
         );
         let (kind, error_context, rust_name) = if let Some(trait_details) = trait_details {
             trait_details
-        } else if let Some(self_ty) = self_ty {
+        } else if let Some(ref self_ty) = self_ty {
             // Some kind of method or static method.
             let type_ident = self_ty.get_final_item();
             // bindgen generates methods with the name:
@@ -837,7 +839,7 @@ impl<'a> FnAnalyzer<'a> {
             let mut rust_name = ideal_rust_name;
             let nested_type_ident = self
                 .nested_type_name_map
-                .get(&self_ty)
+                .get(self_ty)
                 .map(|s| s.as_str())
                 .unwrap_or_else(|| self_ty.get_final_item());
             if matches!(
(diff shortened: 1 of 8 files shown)


adetaylor (commented):
This feels very much along the right lines. In fact you've gone further than I thought you'd do in the initial iteration by tackling the MoveIt TryNew stuff.

Before I land this, I'd want to see:
* Diagnostics if a `throws!` directive had no effect
* Documentation, obviously
* All existing tests passing (I'll click the Approve and Run button shortly)

Generally speaking this is looking great.

torrancew:
Thanks!

The test suite is green locally, and the failing tests appear to be the ones I've added. I'm inclined to suspect a difference in my build environment and CI's. I'll dig into that and try to make them a bit more portable, assuming it's correct. Likewise I'll start in on cleaning up the docs. In the mean time, I added a few more test cases, and in the process uncovered (and fixed) support for constructors as well!

engine/src/conversion/analysis/fun/mod.rs
```diff
@@ -1196,6 +1198,16 @@ impl<'a> FnAnalyzer<'a> {
         }
         let mut cxxbridge_name = make_ident(&cxxbridge_name);
 
+        let friendly_name = {
```
adetaylor, line 1201: I appreciate that much of the naming stuff is a bit of a mess, but please do your best to be consistent here with the way other configuration directives are approached.

In `byvalue_checker.rs` we have:

```rust
        let pod_requests = config
            .get_pod_requests()
            .iter()
            .map(|ty| QualifiedName::new_from_cpp_name(ty))
            .collect();
```

and the comment in `new_from_cpp_name` suggests this is what you should be using.

engine/src/conversion/codegen_rs/fun_codegen.rs
```diff
@@ -196,6 +198,16 @@ pub(super) fn gen_function(
     // which the user has declared.
     let params = unqualify_params(params);
     let ret_type = unqualify_ret_type(ret_type.into_owned());
+
+    let ret_type = if analysis.throwable {
```
adetaylor, line 202: Slight preference for a single `match` statement with an arm something like `_ if !analysis.throwable => ret_type`

adetaylor:
@torrancew Hi, I've been doing a thorough cleanup of all 27 outstanding autocxx PRs so that we don't get to the point of PR bankruptcy. Yours is the *only* one I haven't ruthlessly closed or merged. It would be good to see it get wrapped up!

torrancew:
Thanks for the bump! I'll try to make some time to clean up the diagnostics
and feed back this weekend!

-Tray

PR #1478, opened 2025-03-17
Support POD types with bitfields

tchebb:
Although bindgen has [supported bitfields](https://rust-lang.github.io/rust-bindgen/using-bitfields.html) for a long time, autocxx [currently forbids them](https://github.com/google/autocxx/pull/1313) in POD types (issue #1303). The main obstacle to support is the way bindgen represents bitfields, using a custom struct (emitted alongside the bindings) named `__BindgenBitfieldUnit<T>`, where `T` is the underlying storage type (e.g. `[u8; 4usize]` for a `uint32_t` bitfield).

Because bitfields are structs on the Rust side, autocxx wants `__BindgenBitfieldUnit` to be POD before it'll let any struct that includes one be POD. But there are a few reasons we shouldn't/can't mark `__BindgenBitfieldUnit` as a POD type:

1. It's not semantically correct: POD is a property of types from C++, but `__BindgenBitfieldUnit` is purely a Rust-side abstraction.
2. The double underscore at the beginning of `__BindgenBitfieldUnit` causes our `validate_ident_ok_for_cxx()` to reject it, meaning it never gets added to `apis` when we parse bindgen's output.
3. For the same reason, cxx rejects `__BindgenBitfieldUnit` as a type in the bridge module.
4. Even ignoring all that, our existing struct analysis can't handle structs with type parameters at all, and our `convert_type()` can't handle array-type type parameters.

We can't easily fix these issues, but we can sidestep them! Like point 1 says, it doesn't make sense to treat `__BindgenBitfieldUnit` as a C++ type at all. So instead of trying to analyze it as one, we can treat it as what it is—a primitive type that happens to have a somewhat odd spelling on the Rust side.

This commit adds a few special cases to make `analyze_pod_apis()` "see through" fields of type `__BindgenBitfieldUnit`, treating them like the underlying storage type (taken from the type parameter) instead of like structs. And that's all we need: the (now-deleted) error message that implies cxx can't handle types with `__BindgenBitfieldUnit` members turns out to be wrong: cxx [doesn't care at all](https://github.com/dtolnay/cxx/blob/33084bae28e7b690f977b27d7bdab11f3bab8ce4/syntax/check.rs#L318-L354) about the names of fields inside ExternType structs, so it needs no changes.

Interestingly, we unwrap the bitfield type in exactly the same way we unwrap the bindgen marker types from #1442. We even use the same code!

This PR depends on adetaylor/rust-bindgen#15. Without it, things will *mostly* work, but some types that end with bitfields (including the one in the new integration tests) will fail to generate due to the opaque padding field.

If you merge this PR, please don't squash it. The commits are already split appropriately.

### DIFF
--- a/engine/src/lib.rs
+++ b/engine/src/lib.rs
@@ -345,6 +345,7 @@ impl IncludeCppEngine {
             .clang_args(make_clang_args(inc_dirs, extra_clang_args))
             .derive_copy(false)
             .derive_debug(false)
+            .derive_default(true) // Needed to safely construct POD structs with bitfields
             .default_enum_style(bindgen::EnumVariation::Rust {
                 non_exhaustive: false,
             })
--- a/engine/src/types.rs
+++ b/engine/src/types.rs
@@ -236,10 +236,8 @@ impl std::fmt::Debug for QualifiedName {
 /// cxx.
 #[derive(Error, Clone, Debug)]
 pub enum InvalidIdentError {
-    #[error("Union are not supported by autocxx (and their bindgen names have __ so are not acceptable to cxx)")]
+    #[error("Union are not supported by autocxx")]
     Union,
-    #[error("Bitfields are not supported by autocxx (and their bindgen names have __ so are not acceptable to cxx)")]
-    Bitfield,
     #[error("Names containing __ are reserved by C++ so not acceptable to cxx")]
     TooManyUnderscores,
     #[error("bindgen decided to call this type _bindgen_ty_N because it couldn't deduce the correct name for it. That means we can't generate C++ bindings to it.")]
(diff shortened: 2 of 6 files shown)


engine/src/lib.rs
```diff
@@ -345,6 +345,7 @@ impl IncludeCppEngine {
             .clang_args(make_clang_args(inc_dirs, extra_clang_args))
             .derive_copy(false)
             .derive_debug(false)
+            .derive_default(true) // Needed to safely construct POD structs with bitfields
```
tchebb, line 348: Let me know if you think this'll have detrimental code size or correctness effects elsewhere. As per [bindgen's documentation](https://rust-lang.github.io/rust-bindgen/using-bitfields.html), it's the only ergonomic way to construct types with bitfields, but I can leave it out if it breaks other things.

engine/src/types.rs
```diff
@@ -236,10 +236,8 @@ impl std::fmt::Debug for QualifiedName {
 /// cxx.
 #[derive(Error, Clone, Debug)]
 pub enum InvalidIdentError {
-    #[error("Union are not supported by autocxx (and their bindgen names have __ so are not acceptable to cxx)")]
+    #[error("Union are not supported by autocxx")]
     Union,
-    #[error("Bitfields are not supported by autocxx (and their bindgen names have __ so are not acceptable to cxx)")]
-    Bitfield,
```
tchebb, line 242: I can leave this variant in (and just never construct it) if you don't want a semver break. Let me know.

